PR #56, opened 2025-04-10
integrate error variable of KAL

sivchari:
Some same error variable are defined multiple times and these variable will be defined in the future.
So let's integrate into one as errors.


JoelSpeed (commented):
Makes sense to me, thanks!
/lgtm
/approve

PR #134, opened 2025-08-05
Ignore field that has schemaless marker

### LINKED ISSUE #120: KAL recommends markers on Schemaless fields
Finding
```
api/core/v1beta2/clusterclass_types.go:792:2: maxlength: field AllOf must have a maximum items, add kubebuilder:validation:MaxItems marker (kubeapilinter)
        AllOf []JSONSchemaProps `json:"allOf,omitempty"`
        ^

api/core/v1beta2/clusterclass_types.go:792:2: ssatags: AllOf should have a listType marker for proper Server-Side Apply behavior (atomic, set, or map) (kubeapilinter)
        AllOf []JSONSchemaProps `json:"allOf,omitempty"`
        ^
```

Types
```go
	// allOf specifies that the variable must validate against all of the subschemas in the array.
	// NOTE: This field uses PreserveUnknownFields and Schemaless,
	// because recursive validation is not possible.
	// +optional
	// +kubebuilder:pruning:PreserveUnknownFields
	// +kubebuilder:validation:Schemaless
	AllOf []JSONSchemaProps `json:"allOf,omitempty"`
```

Field in CRD YAML:
```yaml
                                  allOf:
                                    description: |-
                                      allOf specifies that the variable must validate against all of the subschemas in the array.
                                      NOTE: This field uses PreserveUnknownFields and Schemaless,
                                      because recursive validation is not possible.
                                    x-kubernetes-preserve-unknown-fields: true
```

If I add the markers to the field I get this error from controller-gen
```
api/core/v1beta2/clusterclass_types.go:793:2: must apply maxitem to an array
/api/core/v1beta2/clusterclass_types.go:794:2: must apply listType to an array, found
```

I think controller-gen is right that if a field is Schemaless we shouldn't add markers to extend the schema

(We have to use Schemaless in this case because otherwise we would end up with a recursive schema)

sivchari:
close #120

### DIFF
--- a/pkg/analysis/helpers/inspector/inspector.go
+++ b/pkg/analysis/helpers/inspector/inspector.go
@@ -66,65 +66,69 @@ func (i *inspector) InspectFields(inspectField func(field *ast.Field, stack []as
 			return false
 		}
 
-		if len(stack) < 3 {
-			return true
-		}
-
-		// The 0th node in the stack is the *ast.File.
-		// The 1st node in the stack is the *ast.GenDecl.
-		decl, ok := stack[1].(*ast.GenDecl)
-		if !ok {
-			// Make sure that we don't inspect structs within a function.
-			return false
-		}
-
-		if decl.Tok != token.TYPE {
-			// Returning false here means we won't inspect non-type declarations (e.g. var, const, import).
-			return false
-		}
-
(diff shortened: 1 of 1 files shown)


sivchari:
@JoelSpeed 
Sorry for late response. I addressed it.

PR #159, opened 2025-10-24
Minor changes to function definition for consitency

Karthik-K-N:
While exploring on https://github.com/kubernetes-sigs/kube-api-linter/issues/158 found these things, Made changes to keep consistency across all the files.

Also I see that argument `[]ast.Node` of `InspectFields` method is never used. Any reason for keeping that?


Karthik-K-N:
Got the points, I will clean it up to use separate function and remove unused arg. Thanks

PR #166, opened 2025-10-28
arrayofstruct: follow-up on review

saschagrunert:
- Updated hasRequiredField to use the exported utility
- Added recursion exit condition for basic types
- Updated error message to avoid specifying exact marker

Follow-up on https://github.com/kubernetes-sigs/kube-api-linter/pull/162

### DIFF
--- a/pkg/analysis/arrayofstruct/analyzer.go
+++ b/pkg/analysis/arrayofstruct/analyzer.go
@@ -25,7 +25,6 @@ import (
 	"sigs.k8s.io/kube-api-linter/pkg/analysis/helpers/inspector"
 	markershelper "sigs.k8s.io/kube-api-linter/pkg/analysis/helpers/markers"
 	"sigs.k8s.io/kube-api-linter/pkg/analysis/utils"
-	"sigs.k8s.io/kube-api-linter/pkg/markers"
 )
 
 const name = "arrayofstruct"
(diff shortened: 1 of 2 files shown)


pkg/analysis/arrayofstruct/analyzer.go
```diff
@@ -119,7 +118,7 @@ func reportArrayOfStructIssue(pass *analysis.Pass, field *ast.Field) {
 		prefix = fieldName
 	}
 
-	message := fmt.Sprintf("%s is an array of structs, but the struct has no required fields. At least one field should be marked as %s to prevent ambiguous YAML configurations", prefix, markers.RequiredMarker)
+	message := fmt.Sprintf("%s is an array of structs, but the struct has no required fields. At least one field should be marked as required to prevent ambiguous YAML configurations", prefix)
```
everettraven: Do we need test updates with this change? I'd suspect we do.

sivchari: +1

sivchari: I don't think this change is enough because I can't see this message as `want` in test-cases yet. You just added valid case, but we also want invalid case of this message.

pkg/analysis/utils/zero_value.go
```diff
@@ -434,9 +434,9 @@ func isFloatIdent(ident *ast.Ident) bool {
...
 }
 
-// isFieldRequired checks if the field is required.
+// IsFieldRequired checks if the field is required.
 // It checks for the presence of the required marker, the kubebuilder required marker, or the k8s required marker.
-func isFieldRequired(field *ast.Field, markersAccess markershelper.Markers) bool {
+func IsFieldRequired(field *ast.Field, markersAccess markershelper.Markers) bool {
 	fieldMarkers := markersAccess.FieldMarkers(field)
 
 	return fieldMarkers.Has(markers.RequiredMarker) ||
```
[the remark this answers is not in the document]
saschagrunert: Yep, rebased and the change is already included now.

PR #174, opened 2025-10-31
Print qualified field name in error messages

Karthik-K-N:
Fixes: https://github.com/kubernetes-sigs/kube-api-linter/issues/158

### DIFF
--- a/pkg/analysis/uniquemarkers/testdata/src/a/k8s.go
+++ b/pkg/analysis/uniquemarkers/testdata/src/a/k8s.go
@@ -41,5 +41,5 @@ type K8s struct {
 
 	// +k8s:listType=map
 	// +k8s:listType=atomic
-	NonUniqueListType []string // want "field NonUniqueListType has multiple definitions of marker k8s:listType when only a single definition should exist"
+	NonUniqueListType []string // want "field K8s.NonUniqueListType has multiple definitions of marker k8s:listType when only a single definition should exist"
 }
--- a/tests/integration/testdata/default_configurations/affinity.go
+++ b/tests/integration/testdata/default_configurations/affinity.go
@@ -2,13 +2,13 @@ package defaultconfigurations
 
 // Affinity is a group of affinity scheduling rules.
 type Affinity struct {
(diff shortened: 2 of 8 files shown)


pkg/analysis/helpers/inspector/inspector.go
```diff
@@ -117,7 +132,7 @@ func (i *inspector) shouldSkipField(field *ast.Field) bool {
 }
 
 // processFieldWithRecovery processes a field with panic recovery.
-func (i *inspector) processFieldWithRecovery(field *ast.Field, inspectField func(field *ast.Field, jsonTagInfo extractjsontags.FieldTagInfo, markersAccess markers.Markers)) {
+func (i *inspector) processFieldWithRecovery(field *ast.Field, qualifiedFieldName string, inspectField func(field *ast.Field, jsonTagInfo extractjsontags.FieldTagInfo, markersAccess markers.Markers, qualifiedFieldName string)) {
```
krishagarwal278: Hi @Karthik-K-N was wondering if there is a specific reason to pass qualifiedFieldName string twice here?

JoelSpeed (commented):
Approach LGTM, please continue

pkg/analysis/helpers/inspector/inspector.go
Karthik-K-N: Thats because `qualifiedFieldName` is constructed in `i.inspector.WithStack() `function and passed to `processFieldWithRecovery` function and from where it will be passed to the `inspectField` function.

Function definition of `inspectField` function looks like this
```
inspectField func(field *ast.Field, jsonTagInfo extractjsontags.FieldTagInfo, markersAccess markers.Markers, qualifiedFieldName string)
```

inspectField is what used by the analyzers.

Karthik-K-N:
@JoelSpeed PR is ready for review, Please take a look when time permits, Thanks.

JoelSpeed (commented):
Apart from the merge conflicts, this LGTM, thanks @Karthik-K-N

PR #177, opened 2025-11-03
Skip statussubresource linting for Kubernetes List types

### LINKED ISSUE #53: markers helper can't detect markers which are separated with space.
https://github.com/kubernetes-sigs/cluster-api/blob/main/api/v1beta2/cluster_types.go#L1285-L1296

```go
// +kubebuilder:object:root=true

// ClusterList contains a list of Cluster.
type ClusterList struct {
	metav1.TypeMeta `json:",inline"`
	// metadata is the standard list's metadata.
	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#lists-and-simple-kinds
	// +optional
	metav1.ListMeta `json:"metadata,omitempty"`
	// items is the list of Clusters.
	Items []Cluster `json:"items"`
}
```

Above code must be reported by statussubresource linter since this custom resource has root marker and must also have status marker.
But currently marker accessor helper can't collect it because there is space between root marker and ClusterList type.

So I think we should discuss about 2 topic.

1. The way to be able to collect all markers in spite of space.  
2. Configuration for statussubresource. (e.g. ignore error when the type has List suffix on itself.)

saschagrunert:
The `statussubresource` linter was incorrectly flagging Kubernetes List types that don't have a status field. List types follow a different pattern (TypeMeta, ListMeta, Items) and don't use the status subresource.

Fixes #53

### DIFF
--- a/pkg/analysis/helpers/inspector/inspector.go
+++ b/pkg/analysis/helpers/inspector/inspector.go
@@ -117,7 +117,7 @@ func (i *inspector) shouldProcessField(stack []ast.Node, skipListTypes bool) boo
 		return false
 	}
 
-	if skipListTypes && isItemsType(structType) {
+	if skipListTypes && utils.IsKubernetesListType(structType, "") {
 		// Skip list types if requested.
 		return false
 	}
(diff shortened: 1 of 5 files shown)


everettraven (approved):
Aside from one, non-blocking, question about the `nolint:...` usage, this looks good to me.

/lgtm

everettraven:
Will need @JoelSpeed for proper approval

PR #207, opened 2025-12-22
Add table explaining defaulting and scope of linters to docs

### LINKED ISSUE #204: Standardize documentation for reporting linter enabled/disabled defaults
In what I believe to be official linter documentation [1], it has been challenging for me to get an accurate understanding of which linters are enabled by default and which are not. Some linter texts clearly indicate whether they are enabled or disabled by default, but most do not indicate either way.

It would be beneficial for me and other users to standardize how default enabled/disabled is reported in the docs and to make sure that the information is kept up-to-date.

[1] https://github.com/kubernetes-sigs/kube-api-linter/blob/main/docs/linters.md

JoelSpeed:
Fixes #204 

This introduces a table, rather than a list of linters, including additional information in the docs for whether a linter rule is enabled by default and whether it is applicable for Native, CRD or both

### DIFF
--- a/docs/linters.md
+++ b/docs/linters.md
@@ -1,35 +1,39 @@
 # Linters
 
-- [ArrayOfStruct](#arrayofstruct) - Ensures arrays of structs have at least one required field
-- [Conditions](#conditions) - Checks that `Conditions` fields are correctly formatted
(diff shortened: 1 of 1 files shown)


sivchari:
/lgtm
/hold

We want to get feedback from @BlaineEXE

PR #219, opened 2026-01-29
Fix optionalfields linter to handle named types from external packages

### LINKED ISSUE #218: `optionalfields` linter should ignore named map and slice types
### Description

The `optionalfields` linter currently requires all fields marked with `// +optional` to be pointers (e.g., `*int32`). This ensures that "zero values" do not conflict with "unset" values during JSON serialization (using `omitempty`).

However, in Go, maps and slices are reference types. A nil map or slice is already distinguishable from an empty one and is correctly omitted during serialization. Requiring a pointer to a map or a slice is redundant and non-idiomatic Go.

The linter correctly ignores raw maps (e.g., `map[string]string`) and slices. However, it fails to recognize Named Types whose underlying type is a map or slice.This was discovered while enabling the `optionalfields` rule across k8s main repo [PR #135931](https://github.com/kubernetes/kubernetes/pull/135931).

### Example/How to Reproduce

In `k8s.io/api/node/v1/types.go`, the Overhead struct contains:

```
type Overhead struct {
    // +optional
    PodFixed corev1.ResourceList `json:"podFixed,omitempty"`
}
```

Even though `corev1.ResourceList` is defined as `map[ResourceName]resource.Quantity` , the linter triggers the following error: `optionalfields: field Overhead.PodFixed should be a pointer.`

mandre:
Fixes the `optionalfields` linter to correctly handle named types from external packages (e.g., `corev1.ResourceList`). Previously, the linter couldn't resolve underlying types through AST lookups for external packages, causing it to incorrectly require pointers for fields using external map/slice types. This fix extends `IsPointerType`, `IsZeroValueValid`, `GetTypedZeroValue`, `GetTypedValidationHint`, and `hasExplicitZeroMinValidation` to handle `*ast.SelectorExpr` by using Go's type system, enabling correct pointer requirements and accurate error messages for external maps, slices, basic types, and structs.

Assisted-by: claude-opus-4-5@20251101

Fixes #218


JoelSpeed:
/approve
/lgtm

Apologies for the delay in reviewing this

PR #221, opened 2026-02-05
extend maxlength linter to support the declarative validation markers

itzPranshul:
**What This PR does ?**

This PR extends the `maxlength` linter to support k8s declarative validation markers. 

The following markers are recognized-
- `+k8s:maxlength` - on string fields .
- `+k8s:maxItems` - on slice fields.
- `+k8s:enum`- fields and types marked with this are correctly exempted from max length checks.

### DIFF
--- a/pkg/analysis/doc.go
+++ b/pkg/analysis/doc.go
@@ -46,6 +46,10 @@ Example:
 				PreferredOptionalMarker: optionalorrequired.OptionalMarker,
 				PreferredRequiredMarker: optionalorrequired.RequiredMarker,
 			},
+			MaxLength: config.MaxLengthConfig{
+				PreferredMaxLengthMarker: markers.KubebuilderMaxLengthMarker,
+				PreferredMaxItemsMarker:  markers.KubebuilderMaxItemsMarker,
+			},
 		},
 	)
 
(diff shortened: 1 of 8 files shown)


sivchari:
@JoelSpeed In the `defaults` linter, `+k8s:default` alone is not
  sufficient - it warns to add the kubebuilder marker as well (for
  k/k transition).

  Should `maxlength` follow the same approach, or is it OK to treat
  k8s markers as alternatives here?

pkg/analysis/maxlength/testdata/src/a/c.go
```diff
@@ -0,0 +1,21 @@
...
+type K8sMarkers struct {
+	// +k8s:maxLength=256
+	StringWithK8sMaxLength string
+
+	// +k8s:maxItems=256
+	ArrayWithK8sMaxItems []int
+
+	// This field has NO markers, so it SHOULD trigger a diagnostic.
+	// This confirms the linter is running on this file and checks for missing markers.
+	StringWithoutAnyMarkers string // want "field K8sMarkers.StringWithoutAnyMarkers must have a maximum length, add kubebuilder:validation:MaxLength marker"
```
JoelSpeed: We should make sure that this preferred marker is configurable, as we do in for example the optionalorrequired linter implementation

itzPranshul: Done! I have updated the linter to use configrable initializer , similar to `optionalorrequired` .

Introduced a `MaxLengthConfig` struct that allows configuring preferred markers. It defaults to use kubebuilder markers but allows passing new DV tags (`k8s:maxLength` and `k8s:maxItems`) as alternatives.  Also, I've added a dedicated test suite (`testdata/src/b`) to verify the behavior.

itzPranshul:
@JoelSpeed should we extend `maxLength` to support all "maximum" bounds (e.g. `maxProperties`,` maximum`, `maxBytes` ) to ensure consistent resource bounding across API types? Also, if we broaden the scope, would you consider renaming the linter to something more generic like `limits` or `max-bounds` ?

JoelSpeed:
> should we extend maxLength to support all "maximum" bounds (e.g. maxProperties, maximum, maxBytes )

Yes, anything that is being implemented in DV that would make sense here seems reasonable to me

> Also, if we broaden the scope, would you consider renaming the linter to something more generic like limits or max-bounds ?

It all feels like it's still imposting maximum lengths/size to data at the end of the day, I think the name is ok for now

PR #230, opened 2026-02-27
Add discriminated unions structure linter

MatteoFari:
## Description

This PR introduces a new `discriminatedunions` linter to enforce structural best practices for discriminated unions.

The new linter, `discriminatedunions`, does the following:

- **Validates Union Structure:** Detects unions from either legacy markers or declarative markers, including field-marker-only unions.
- **Enforces Discriminator/Member Rules:** Requires exactly one discriminator field, requires the discriminator to be marked required, and requires union member fields to be optional (including support for `+unionMember,optional`).
- **Configurable Non-Member Policy:** Enforces or allows non-member fields via config (`nonMemberFields: Forbid | Allow`), with `Forbid` as default.
- **Safe Rollout:** The linter is disabled by default.

This is the first part and intentionally scopes to structure validation only.  
CRD reachability and CEL gating checks will follow.

Part of #20

### DIFF
--- a/pkg/analysis/discriminatedunions/analyzer.go
+++ b/pkg/analysis/discriminatedunions/analyzer.go
@@ -1,0 +1,307 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package discriminatedunions
+
+import (
+	"go/ast"
+	"slices"
+	"strings"
+
+	"golang.org/x/tools/go/analysis"
+
+	kalerrors "sigs.k8s.io/kube-api-linter/pkg/analysis/errors"
+	"sigs.k8s.io/kube-api-linter/pkg/analysis/helpers/extractjsontags"
+	inspectorhelper "sigs.k8s.io/kube-api-linter/pkg/analysis/helpers/inspector"
+	markershelper "sigs.k8s.io/kube-api-linter/pkg/analysis/helpers/markers"
+	"sigs.k8s.io/kube-api-linter/pkg/analysis/utils"
+	markersconsts "sigs.k8s.io/kube-api-linter/pkg/markers"
+)
+
+const name = "discriminatedunions"
+
+func init() {
+	markershelper.DefaultRegistry().Register(
+		markersconsts.UnionMarker,
+		markersconsts.UnionDiscriminatorMarker,
+		markersconsts.UnionMemberMarker,
+		markersconsts.K8sUnionDiscriminatorMarker,
+		markersconsts.K8sUnionMemberMarker,
+		markersconsts.OptionalMarker,
+		markersconsts.KubebuilderOptionalMarker,
+		markersconsts.K8sOptionalMarker,
+		markersconsts.RequiredMarker,
+		markersconsts.KubebuilderRequiredMarker,
+		markersconsts.K8sRequiredMarker,
+	)
+}
+
+type analyzer struct {
+	nonMemberFields NonMemberFieldsPolicy
+}
+
+type unionType struct {
+	typeSpec *ast.TypeSpec
+	name     string
+
+	hasUnionMarker bool
+
+	discriminatorFields []*unionField
+	memberFields        []*unionField
+	nonMemberFields     []*unionField
+}
+
+type unionField struct {
+	field         *ast.Field
+	qualifiedName string
+
+	required bool
+	optional bool
+
+	isDiscriminator      bool
+	isMember             bool
+	memberOptionalMarker bool
+}
+
+type unionFieldClassification struct {
+	isDiscriminator  bool
+	isMember         bool
+	isMemberOptional bool
+}
+
+func newAnalyzer(cfg *Config) *analysis.Analyzer {
+	if cfg == nil {
+		cfg = &Config{}
+	}
+
+	defaultConfig(cfg)
+
+	a := &analyzer{
+		nonMemberFields: cfg.NonMemberFields,
+	}
+
+	return &analysis.Analyzer{
+		Name:     name,
+		Doc:      "Validates discriminated-union marker structure.",
+		Run:      a.run,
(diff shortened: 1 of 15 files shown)


pkg/analysis/discriminatedunions/analyzer.go
```diff
@@ -0,0 +1,318 @@
...
+	value, ok = marker.Arguments[markersconsts.OptionalMarker]
+	if ok {
+		return normalizedMarkerValue(value) == "true"
+	}
+
+	return false
+}
+
+func normalizedMarkerValue(value string) string {
+	return strings.ToLower(strings.TrimSpace(strings.Trim(value, `"'`)))
```
sivchari: I'm not sure why we make value lower.

pkg/analysis/discriminatedunions/config.go
```diff
@@ -0,0 +1,33 @@
...
+const (
+	// NonMemberFieldsForbid reports non-member fields on union types.
+	NonMemberFieldsForbid NonMemberFieldsPolicy = "Forbid"
+	// NonMemberFieldsAllow allows non-member fields on union types.
+	NonMemberFieldsAllow NonMemberFieldsPolicy = "Allow"
+)
+
+// Config contains the configuration for the discriminatedunions linter.
+type Config struct {
+	// NonMemberFields defines how fields that are neither discriminator nor member are handled.
```
JoelSpeed, line 32: Please expand the comment here to explain the valid values and what they mean to a person configuring this project

pkg/analysis/discriminatedunions/analyzer.go
```diff
@@ -0,0 +1,307 @@
...
+
+type unionType struct {
+	typeSpec *ast.TypeSpec
+	name     string
+
+	hasUnionMarker bool
+
+	discriminatorFields []*unionField
+	memberFields        []*unionField
+	nonMemberFields     []*unionField
```
JoelSpeed: What's the motivation for these being slices of pointers vs slices of references?

pkg/analysis/discriminatedunions/analyzer.go
```diff
@@ -0,0 +1,307 @@
...
+
+	defaultConfig(cfg)
+
+	a := &analyzer{
+		nonMemberFields: cfg.NonMemberFields,
+	}
+
+	return &analysis.Analyzer{
+		Name:     name,
+		Doc:      "Validates discriminated-union marker structure.",
```
JoelSpeed: PTAL at other doc lines from other analyzers, they typically include more information describing the behaviour to the end user

pkg/analysis/discriminatedunions/analyzer.go
```diff
@@ -0,0 +1,307 @@
...
+	if !ok || structType.Fields == nil {
+		return nil
+	}
+
+	structMarkers := markersAccess.StructMarkers(structType)
+
+	union := &unionType{
+		typeSpec:       typeSpec,
+		name:           typeSpec.Name.Name,
+		hasUnionMarker: structMarkers.Has(markersconsts.UnionMarker),
```
JoelSpeed: Or the k8s equivalent marker?

pkg/analysis/discriminatedunions/analyzer.go
```diff
@@ -0,0 +1,307 @@
...
+	for _, field := range structType.Fields.List {
+		unionFieldInfo, ok := fieldInfos[field]
+		if !ok {
+			continue
+		}
+
+		addUnionField(union, unionFieldInfo)
+	}
+
+	if !union.hasUnionMarker && len(union.discriminatorFields) == 0 && len(union.memberFields) == 0 {
```
JoelSpeed: Why does no discriminator get a pass? Why is zero members skipped?

pkg/analysis/discriminatedunions/analyzer.go
```diff
@@ -0,0 +1,307 @@
...
+		union.nonMemberFields = append(union.nonMemberFields, field)
+	}
+}
+
+func reportDiscriminatorViolations(pass *analysis.Pass, union *unionType) {
+	switch len(union.discriminatorFields) {
+	case 0:
+		pass.Reportf(
+			union.typeSpec.Pos(),
+			"type %s is marked as a discriminated union but has no discriminator field; expected exactly one field with +%s or +%s",
```
JoelSpeed: In other linters, we allow a choice of the preferred marker so that the output only contains a single recommended marker. Can we do the same here please?

pkg/analysis/discriminatedunions/analyzer.go
```diff
@@ -0,0 +1,307 @@
...
+			len(union.discriminatorFields),
+			strings.Join(discriminatorNames, ", "),
+		)
+	}
+}
+
+func reportMissingMemberViolations(pass *analysis.Pass, union *unionType) {
+	if len(union.memberFields) == 0 {
+		pass.Reportf(union.typeSpec.Pos(), "type %s is marked as a discriminated union but has no union member fields", union.name)
+	}
```
JoelSpeed: This might not actually be an issue. A union with only a discriminator could well be valid as a set up for a later expansion to include union members that need a member

pkg/analysis/discriminatedunions/analyzer.go
```diff
@@ -0,0 +1,307 @@
...
+}
+
+func reportMemberOptionalityViolations(pass *analysis.Pass, union *unionType) {
+	for _, member := range union.memberFields {
+		if member.optional || member.memberOptionalMarker {
```
JoelSpeed: These don't mean the same thing. All union members require `member.optional`. The `member.memberOptionalMarker` is a separate thing that determines if the member is required when the discriminator has the members equivalent value set

JoelSpeed (commented):
How would we feel about optionally also enforcing CEL validations for the union membership? 

There are [two patterns](https://github.com/openshift/api/blob/ef71f942ef6cca617c5ec7344ec6813949232597/example/v1/types_stable.go#L140-L141) we use in openshift, one for required members, one for optional members. They are tied only to the discriminator value and then the field name itself, so scale well.

The linter could automatically detect these/configure these for users possibly?

pkg/analysis/discriminatedunions/analyzer.go
MatteoFari: I thought every union needed a member but discriminator-only unions are valid, so I removed that check.

MatteoFari:
> How would we feel about optionally also enforcing CEL validations for the union membership?
> 
> There are [two patterns](https://github.com/openshift/api/blob/ef71f942ef6cca617c5ec7344ec6813949232597/example/v1/types_stable.go#L140-L141) we use in openshift, one for required members, one for optional members. They are tied only to the discriminator value and then the field name itself, so scale well.
> 
> The linter could automatically detect these/configure these for users possibly?

I originally thought about this PR as a first part and later add CRD reachability and CEL gating checks. Do you prefer that I include them in this PR?

PR #238, opened 2026-04-07
fix(requiredfields,optionalfields): account for ExactlyOneOf/AtLeastOneOf in zero value detection

### LINKED ISSUE #236: requiredfields: not satisified by `ExactlyOneOf`/`AtLeastOneOf`
The requiredfields analysis does not seem to take into account the struct-level marker `+kubebuilder:validation:ExactlyOneOf` when determining if a struct's zero value is invalid. In my project it thus requires a field be turned into a pointer unnecessarily.

```go
// OriginIssuerSpec is the specification of an OriginIssuer. This includes any
// configuration required for the issuer.
type OriginIssuerSpec struct {
	// requestType is the signature algorithm Cloudflare should use to sign the certificate.
	// +required
	RequestType RequestType `json:"requestType,omitempty"`

	// auth configures how to authenticate with the Cloudflare API.
	// +required
	Auth OriginIssuerAuthentication `json:"auth,omitzero"`
}

// OriginIssuerAuthentication defines how to authenticate with the Cloudflare API.
// Only one of `serviceKeyRef` or `tokenRef` may be specified.
//
// +kubebuilder:validation:ExactlyOneOf=serviceKeyRef;tokenRef
type OriginIssuerAuthentication struct {
	// serviceKeyRef authenticates with an API Service Key (the "Origin CA Key").
	// +optional
	ServiceKeyRef *SecretKeySelector `json:"serviceKeyRef,omitempty"`

	// tokenRef authenticates with an API Token.
	// +optional
	TokenRef *SecretKeySelector `json:"tokenRef,omitempty"`
}
```

> requiredfields: field OriginIssuerSpec.Auth has a valid zero value ({}), but the validation is not complete (e.g. min properties/adding required fields). The field should be a pointer to allow the zero value to be set. If the zero value is not a valid use case, complete the validation and remove the pointer.

This seems similar to #178.

venkatasai-kadamati:
## Summary

The `requiredfields` and `optionalfields` analyzers did not consider struct-level `ExactlyOneOf`/`AtLeastOneOf` markers when determining if a struct's zero value is valid. These markers enforce that at least one field must be set, making `{}` an invalid zero value.

## Design

Instead of adding a parallel conditional branch, the fix unifies union markers with the existing `minProperties` flow via two helpers:

- **`hasUnionConstraint(markerSet)`** — single predicate for all union markers; the sole extension point for future markers (e.g. `MaxOneOf`).
- **`resolveEffectiveMinProperties(markerSet)`** — synthesizes an effective `minProperties=1` when union markers are present and no explicit `minProperties` is set. Delegates to `GetMinProperties`; union constraint is applied only when explicit min is absent.

The downstream logic in `isStructZeroValueValid` is unchanged — it now operates on `effectiveMin` instead of `minProperties` directly.

## Changes

- **`pkg/analysis/utils/zero_value.go`**: Extract `hasUnionConstraint` and `resolveEffectiveMinProperties`; refactor `isStructZeroValueValid` to use `effectiveMin`; update godoc.
- **`pkg/analysis/requiredfields/analyzer.go`**: Register the two new markers in `init()`.
- **`pkg/analysis/optionalfields/analyzer.go`**: Register the two new markers in `init()`.
- **Test data — `requiredfields`**: 4 field variants (plain, omitempty, pointer, pointer+omitempty) for both markers; test for `nonOmittedFields>0` case (ExactlyOneOf + required field, union check correctly silent); test for `MinProperties=1` + `ExactlyOneOf` coexistence.
- **Test data — `optionalfields`**: 4 field variants for both markers; `+required` field regression case.

## Naming note

`KubebuilderExactlyOneOf` inconsistently lacks the `Marker` suffix present on `KubebuilderAtLeastOneOfMarker`. A `TODO` is added in `hasUnionConstraint`; the rename is tracked separately to avoid an unrelated change in this PR.

## Test plan

- [x] `go test ./pkg/analysis/requiredfields/` — pass
- [x] `go test ./pkg/analysis/optionalfields/` — pass
- [x] `go test ./pkg/analysis/utils/...` — pass
- [x] `go test ./...` — all 40 analysis packages pass

Fixes kubernetes-sigs/kube-api-linter/issues/236

### DIFF
--- a/pkg/analysis/requiredfields/testdata/src/a/structs.go
+++ b/pkg/analysis/requiredfields/testdata/src/a/structs.go
@@ -84,6 +84,53 @@ type TestStructs struct {
 
 	// +required
 	StructPtrWithOmittedRequiredFieldWithOmitEmpty *StructWithOmittedRequiredField `json:"structPtrWithOmittedRequiredFieldWithOmitEmpty,omitempty"` // want "field TestStructs.StructPtrWithOmittedRequiredFieldWithOmitEmpty does not allow the zero value. It must have the omitzero tag." "field TestStructs.StructPtrWithOmittedRequiredFieldWithOmitEmpty does not allow the zero value. The field does not need to be a pointer."
+
+	// StructWithExactlyOneOf has a zero value of {}, which is not valid because the ExactlyOneOf marker requires exactly one field to be set.
+
+	// +required
+	StructWithExactlyOneOf StructWithExactlyOneOf `json:"structWithExactlyOneOf"` // want "field TestStructs.StructWithExactlyOneOf does not allow the zero value. It must have the omitzero tag."
+
+	// +required
+	StructWithExactlyOneOfWithOmitEmpty StructWithExactlyOneOf `json:"structWithExactlyOneOfWithOmitEmpty,omitempty"` // want "field TestStructs.StructWithExactlyOneOfWithOmitEmpty does not allow the zero value. It must have the omitzero tag."
+
+	// +required
+	StructPtrWithExactlyOneOf *StructWithExactlyOneOf `json:"structPtrWithExactlyOneOf"` // want "field TestStructs.StructPtrWithExactlyOneOf does not allow the zero value. It must have the omitzero tag." "field TestStructs.StructPtrWithExactlyOneOf does not allow the zero value. The field does not need to be a pointer."
+
+	// +required
+	StructPtrWithExactlyOneOfWithOmitEmpty *StructWithExactlyOneOf `json:"structPtrWithExactlyOneOfWithOmitEmpty,omitempty"` // want "field TestStructs.StructPtrWithExactlyOneOfWithOmitEmpty does not allow the zero value. It must have the omitzero tag." "field TestStructs.StructPtrWithExactlyOneOfWithOmitEmpty does not allow the zero value. The field does not need to be a pointer."
+
+	// StructWithAtLeastOneOf has a zero value of {}, which is not valid because the AtLeastOneOf marker requires at least one field to be set.
+
+	// +required
+	StructWithAtLeastOneOf StructWithAtLeastOneOf `json:"structWithAtLeastOneOf"` // want "field TestStructs.StructWithAtLeastOneOf does not allow the zero value. It must have the omitzero tag."
+
+	// +required
+	StructWithAtLeastOneOfWithOmitEmpty StructWithAtLeastOneOf `json:"structWithAtLeastOneOfWithOmitEmpty,omitempty"` // want "field TestStructs.StructWithAtLeastOneOfWithOmitEmpty does not allow the zero value. It must have the omitzero tag."
+
+	// +required
+	StructPtrWithAtLeastOneOf *StructWithAtLeastOneOf `json:"structPtrWithAtLeastOneOf"` // want "field TestStructs.StructPtrWithAtLeastOneOf does not allow the zero value. It must have the omitzero tag." "field TestStructs.StructPtrWithAtLeastOneOf does not allow the zero value. The field does not need to be a pointer."
+
+	// +required
+	StructPtrWithAtLeastOneOfWithOmitEmpty *StructWithAtLeastOneOf `json:"structPtrWithAtLeastOneOfWithOmitEmpty,omitempty"` // want "field TestStructs.StructPtrWithAtLeastOneOfWithOmitEmpty does not allow the zero value. It must have the omitzero tag." "field TestStructs.StructPtrWithAtLeastOneOfWithOmitEmpty does not allow the zero value. The field does not need to be a pointer."
+
+	// StructWithExactlyOneOfOneWithoutOmitEmpty has one union member without omitempty.
+	// That member serializes as null at zero value; has(null) = false in CEL.
+	// ExactlyOneOf is still violated.
+
+	// +required
+	StructWithExactlyOneOfOneWithoutOmitEmpty StructWithExactlyOneOfOneWithoutOmitEmpty `json:"structWithExactlyOneOfOneWithoutOmitEmpty"` // want "field TestStructs.StructWithExactlyOneOfOneWithoutOmitEmpty does not allow the zero value. It must have the omitzero tag."
+
+	// StructWithExactlyOneOfNonUnionNonOmitted has a non-union field without omitempty
+	// alongside union fields. The non-union field does not satisfy ExactlyOneOf.
+
+	// +required
+	StructWithExactlyOneOfNonUnionNonOmitted StructWithExactlyOneOfNonUnionNonOmitted `json:"structWithExactlyOneOfNonUnionNonOmitted"` // want "field TestStructs.StructWithExactlyOneOfNonUnionNonOmitted does not allow the zero value. It must have the omitzero tag."
+
+	// StructWithMinPropertiesAndExactlyOneOf: MinProperties=1 takes precedence in
+	// resolveEffectiveMinProperties; ExactlyOneOf is redundant. Behavior is same as MinProperties=1.
+
+	// +required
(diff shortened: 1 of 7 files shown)


pkg/analysis/utils/zero_value.go
```diff
@@ -91,11 +91,44 @@ func GetTypeMarkerValue(pass *analysis.Pass, field *ast.Field, markersAccess mar
 	return ""
 }
 
+// hasUnionConstraint reports whether the marker set contains a union validation
+// marker that implicitly requires at least one field to be set.
+// It is the single extension point for adding new union markers in the future.
+// hasUnionConstraint is indirectly tested via the requiredfields and optionalfields
+// analyzer integration tests (testdata/src/a/).
+// TODO: rename KubebuilderExactlyOneOf to KubebuilderExactlyOneOfMarker for consistency
+// with KubebuilderAtLeastOneOfMarker (tracked separately).
```
JoelSpeed: Why not do this as a pre-factor commit in this PR?

pkg/analysis/utils/zero_value.go
```diff
@@ -119,21 +152,23 @@ func isStructZeroValueValid(pass *analysis.Pass, field *ast.Field, structType *a
...
 		pass.Reportf(field.Pos(), "struct %s has an invalid minProperties marker: %v", FieldName(field), err)
 		return false, false
 	}
 
-	if minProperties != nil && *minProperties > nonOmittedFields {
-		// The struct requires more properties than would be marshalled in the zero value of the struct.
+	// effectiveMin > nonOmittedFields means the struct's zero value ({} or a subset)
+	// satisfies fewer properties than required. When nonOmittedFields > 0 and
+	// effectiveMin <= nonOmittedFields, the zero-value struct always serializes those
+	// fields and the union constraint is trivially satisfied.
```
JoelSpeed: Not necessarily true. If I had a union across 2 fields, but then another field that wasn't part of the union, but wasn't omitted, then this would say that the zero value is valid, when in fact it's not, which might resolve to the wrong logic.

I think to correctly handle this we need to know whether the union is actually satisfied by determining whether the union members correctly have omitempty

pkg/analysis/utils/zero_value.go
```diff
@@ -119,21 +152,23 @@ func isStructZeroValueValid(pass *analysis.Pass, field *ast.Field, structType *a
...
+	if effectiveMin != nil && *effectiveMin > nonOmittedFields {
 		zeroValueValid = false
 	}
 
 	var completeStructValidation = true
-	if minProperties == nil && nonOmittedFields == 0 {
-		// If the struct has no non-omitted fields, then the zero value of the struct is `{}`.
-		// This generally means that the validation is incomplete as the difference between omitting the field and not omitting is not clear.
```
JoelSpeed, line 136: Maybe retain the `incomplete` part of this as we use `incomplete validation` quite a bit through the project to mean something specific

pkg/analysis/requiredfields/testdata/src/a/structs.go
```diff
@@ -144,3 +184,54 @@ type StructWithOmittedRequiredField struct {
...
+type StructWithExactlyOneOfAndRequiredField struct {
+	// +required
+	RequiredString string `json:"requiredString"` // want "field StructWithExactlyOneOfAndRequiredField.RequiredString has a valid zero value \\(\"\"\\), but the validation is not complete \\(e.g. minimum length\\). The field should be a pointer to allow the zero value to be set. If the zero value is not a valid use case, complete the validation and remove the pointer." "field StructWithExactlyOneOfAndRequiredField.RequiredString should have the omitempty tag."
+
+	// +optional
+	FieldA *string `json:"fieldA,omitempty"`
+
+	// +optional
+	FieldB *string `json:"fieldB,omitempty"`
+}
```
JoelSpeed: What if this struct had a non-omitted optional field alongside the union fields

pkg/analysis/requiredfields/testdata/src/a/structs.go
```diff
@@ -144,3 +184,54 @@ type StructWithOmittedRequiredField struct {
...
+// Struct with ExactlyOneOf marker.
+// The zero value of the struct is `{}` which is not valid because the ExactlyOneOf marker requires exactly one field to be set.
+// +kubebuilder:validation:ExactlyOneOf=serviceKeyRef;tokenRef
+type StructWithExactlyOneOf struct {
+	// +optional
+	ServiceKeyRef *string `json:"serviceKeyRef,omitempty"`
+
+	// +optional
+	TokenRef *string `json:"tokenRef,omitempty"`
+}
```
JoelSpeed, line 203: What about an example where one or both do not have omitempty?

venkatasai-kadamati:
Addressing all review comments — force-pushed with 2 commits as requested.

---

**Re: [Why not do this as a pre-factor commit?](https://github.com/kubernetes-sigs/kube-api-linter/pull/238/files#r3050341983)**

Good call. Restructured into two commits: the first introduces `resolveEffectiveMinProperties` wrapping the existing `GetMinProperties` call with no behavior change (pure preparatory step), the second adds the union constraint logic on top.

---

**Re: [Not necessarily true — nonOmittedFields logic](https://github.com/kubernetes-sigs/kube-api-linter/pull/238/files#r3050356447)**

You are right — the comment and the logic were both wrong. `nonOmittedFields` counts **all** non-omitted struct fields, but `ExactlyOneOf`/`AtLeastOneOf` applies only to the named union members. A non-union required or non-omitted optional field increments `nonOmittedFields` without satisfying the union constraint.

The correct fix is unconditional: per Kubernetes docs, CEL `has(field)` returns false for null values (*"Null valued fields are treated as absent fields in CEL expressions"*). At zero value, all union members are nil (pointer → null) or zero (value type) — neither counts as set. This holds regardless of non-union non-omitted fields.

Replaced with:

```go
switch {
case hasUnionConstraint(markerSet):
    // CEL has(field) = false for null/absent — zero value always violates the constraint.
    zeroValueValid = false
case effectiveMin != nil && *effectiveMin > nonOmittedFields:
    zeroValueValid = false
}
```

---

**Re: [Retain the `incomplete` part](https://github.com/kubernetes-sigs/kube-api-linter/pull/238/files#r3050362380)**

Updated to use "incomplete validation" throughout — consistent with `IsZeroValueValid`'s public godoc and existing diagnostic messages. "Indeterminate" did not appear anywhere else in the codebase.

---

**Re: [Non-omitted optional field alongside union fields](https://github.com/kubernetes-sigs/kube-api-linter/pull/238/files#r3050369763)**

Added `StructWithExactlyOneOfNonUnionNonOmitted` — a struct with a non-omitted optional field (no `+required`, no `omitempty`) alongside union members. This is the exact counterexample to the old logic: `nonOmittedFields=1` from the non-union field incorrectly passed `effectiveMin > nonOmittedFields`, but the union constraint is still violated. With the unconditional fix, this is correctly caught.

---

**Re: [What about an example where one or both do not have omitempty?](https://github.com/kubernetes-sigs/kube-api-linter/pull/238/files#r3050371802)**

Added `StructWithExactlyOneOfOneWithoutOmitEmpty` — a union member without `omitempty` serializes as `{"serviceKeyRef": null}` at zero value. `has(self.serviceKeyRef)` = false (null treated as absent in CEL), so `ExactlyOneOf` is still violated. With the unconditional fix, this is correctly handled regardless of `omitempty` status.

JoelSpeed:
@venkatasai-kadamati could you please confirm if you are, and if so, which AI agent you are using to assist you with this PR?

Per Kubernetes contributor guidelines you are required to disclose any use of AI to assist in your contributions.
It is important for reviewers to understand where AI is being used so that we can adjust our review style accordingly.

venkatasai-kadamati:
Yes, I used Claude to assist with this contribution. Specifically, the AI helped identify the issue, draft the implementation, generate tests, and respond to review comments. I reviewed and approved each step before it was pushed.

I understand the importance of disclosure and apologize for not leading with this. Happy to answer any follow-up questions about the extent of AI involvement.

JoelSpeed:
/close

This has been solved in #237

PR #239, opened 2026-04-07
Fix inline json tag detection

liggitt:
xref https://github.com/kubernetes/kubernetes/pull/138260 https://github.com/kubernetes/kubernetes/issues/138133

1. Flags use of `json:",inline"` with non-embedded fields as problematic
2. Allows use of `json:""` with embedded fields

/cc @JoelSpeed

### DIFF
--- a/pkg/analysis/jsontags/analyzer.go
+++ b/pkg/analysis/jsontags/analyzer.go
@@ -77,8 +77,11 @@ func (a *analyzer) run(pass *analysis.Pass) (any, error) {
 }
 
 func (a *analyzer) checkField(pass *analysis.Pass, field *ast.Field, tagInfo extractjsontags.FieldTagInfo, qualifiedFieldName string) {
+	embedded := false
 	prefix := "field %s"
+
 	if len(field.Names) == 0 || field.Names[0] == nil {
+		embedded = true
 		prefix = "embedded field %s"
 	}
 
@@ -90,11 +93,18 @@ func (a *analyzer) checkField(pass *analysis.Pass, field *ast.Field, tagInfo ext
 	}
 
 	if tagInfo.Inline {
+		if !embedded {
+			pass.Reportf(field.Pos(), "%s has inline json tag, but is not embedded", prefix)
+		}
+
 		return
 	}
 
 	if tagInfo.Name == "" {
-		pass.Reportf(field.Pos(), "%s has empty json tag", prefix)
+		if !embedded {
+			pass.Reportf(field.Pos(), "%s has empty json tag", prefix)
+		}
+
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


liggitt:
uh... none of these linting errors make sense to me...

```
Error: pkg/analysis/jsontags/analyzer.go:98:3: return with no blank line before (nlreturn)
		return
		^
Error: pkg/analysis/jsontags/analyzer.go:105:3: return with no blank line before (nlreturn)
		return
		^
Error: pkg/analysis/jsontags/analyzer.go:82:2: only one cuddle assignment allowed before if statement (wsl)
	if len(field.Names) == 0 || field.Names[0] == nil {
	^
```

JoelSpeed:
For each of the linter errors, just add a single blank line before the line it's complaining about

pkg/analysis/jsontags/analyzer.go
```diff
@@ -77,8 +77,10 @@ func (a *analyzer) run(pass *analysis.Pass) (any, error) {
 }
 
 func (a *analyzer) checkField(pass *analysis.Pass, field *ast.Field, tagInfo extractjsontags.FieldTagInfo, qualifiedFieldName string) {
+	embedded := false
 	prefix := "field %s"
 	if len(field.Names) == 0 || field.Names[0] == nil {
+		embedded = true
 		prefix = "embedded field %s"
 	}
```
JoelSpeed, line 86:
```suggestion
	embedded := false
prefix := "field %s"

if len(field.Names) == 0 || field.Names[0] == nil {
		embedded = true
prefix = "embedded field %s"
}
```

pkg/analysis/jsontags/analyzer.go
```diff
@@ -90,11 +92,16 @@ func (a *analyzer) checkField(pass *analysis.Pass, field *ast.Field, tagInfo ext
...
 
 	if tagInfo.Inline {
+		if !embedded {
+			pass.Reportf(field.Pos(), "%s has inline json tag, but is not embedded", prefix)
+		}
 		return
```
JoelSpeed, line 100:
```suggestion
		if !embedded {
			pass.Reportf(field.Pos(), "%s has inline json tag, but is not embedded", prefix)
		}

return
```

pkg/analysis/jsontags/analyzer.go
```diff
@@ -90,11 +92,16 @@ func (a *analyzer) checkField(pass *analysis.Pass, field *ast.Field, tagInfo ext
...
 	}
 
 	if tagInfo.Name == "" {
-		pass.Reportf(field.Pos(), "%s has empty json tag", prefix)
+		if !embedded {
+			pass.Reportf(field.Pos(), "%s has empty json tag", prefix)
+		}
 		return
```
JoelSpeed, line 108:
```suggestion
		if !embedded {
			pass.Reportf(field.Pos(), "%s has empty json tag", prefix)
		}

return
```

liggitt:
> For each of the linter errors, just add a single blank line before the line it's complaining about

yeah, I know how to fix the linter issue, but are these checks worth having / enforcing?

JoelSpeed:
Up for debate, some like them, some don't, feels the same for any linter. I think the cuddle bit is probably a little too strict but enforcing some whitespace I do like for readability 

Changing the linter config seems orthogonal to the fix though, I'd rather we fix the listing issues now and debate changing the whitespace linter later

JoelSpeed (commented):
/lgtm
/approve

PR #240, opened 2026-04-14
Change golangci-lint version to v2.11.0

emirot:
## What

Update golangci-lint version in README example, so it is more explicit

## Why

I was a bit confused by that because configuration above refer to configuration version and this one is golangci-lint refer to the binary version.


JoelSpeed (commented):
/lgtm
/approve

Thanks!

PR #241, opened 2026-04-14
Fix inline json tag detection

liggitt:
Follow-up to https://github.com/kubernetes-sigs/kube-api-linter/pull/239

Prereq for https://github.com/kubernetes/kubernetes/pull/138260

Part 2 (fixes inline detection so other linters which exempt inline fields work properly)

/assign @JoelSpeed

### DIFF
--- a/pkg/analysis/helpers/extractjsontags/analyzer.go
+++ b/pkg/analysis/helpers/extractjsontags/analyzer.go
@@ -91,14 +91,16 @@ func run(pass *analysis.Pass) (any, error) {
 			return
 		}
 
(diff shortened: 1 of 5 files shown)


liggitt:
```
Error: pkg/analysis/helpers/extractjsontags/analyzer.go:101:6: Function 'extractTagInfo' is too long (62 > 60) (funlen)
func extractTagInfo(field *ast.Field, tag *ast.BasicLit) FieldTagInfo {
```

😒

everettraven (approved):
/lgtm

PR #248, opened 2026-06-01
refactor inspector helper to use iterators instead of callbacks

sivchari:
Introduce iterator pattern instead of callbacks. It's faster and safer than current implementation.

### DIFF
--- a/pkg/analysis/helpers/inspector/inspector.go
+++ b/pkg/analysis/helpers/inspector/inspector.go
@@ -20,6 +20,7 @@ import (
 	"go/ast"
 	"go/token"
 	"go/types"
+	"iter"
 
 	astinspector "golang.org/x/tools/go/ast/inspector"
 	"sigs.k8s.io/kube-api-linter/pkg/analysis/helpers/extractjsontags"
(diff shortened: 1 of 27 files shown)


pkg/analysis/helpers/inspector/inspector.go
```diff
@@ -155,35 +192,37 @@ func (i *inspector) shouldSkipField(field *ast.Field) bool {
...
 	tagInfo := i.jsonTags.FieldTags(field)
 
 	defer func() {
 		if r := recover(); r != nil {
-			// If the inspectField function panics, we recover and log information that will help identify the issue.
+			// If the consumer panics, we recover and log information that will help identify the issue.
 			debug := printDebugInfo(field)
 			panic(fmt.Sprintf("%s %v", debug, r)) // Re-panic to propagate the error.
 		}
 	}()
```
JoelSpeed, line 206: I'm not convinced we will need this recovery now, since the panic is not handled in a closure we call, will this actually do anything?

JoelSpeed (commented):
LGTM, just one question about whether we actually still need that panic recovery at all now

pkg/analysis/helpers/inspector/inspector.go
sivchari: Verified this still works correctly, since yield() is called synchronously within this function's stack frame, the deferred recover() catches panics from the loop body just like before.

sivchari:
Hi @JoelSpeed 
PTAL 🙏

PR #252, opened 2026-07-09
Fix dependenttags value matching for marker dependencies

GaurangKhanderay:
## What this PR does / why we need it

The dependenttags linter currently uses direct MarkerSet key lookups when checking marker dependencies.

This does not correctly handle markers that contain values, such as `+k8s:update=NoUnset`, because marker identifiers and values are stored separately during parsing.

This PR adds a helper function to correctly match value-qualified markers using `MarkerSet.HasWithValue()` while preserving existing behavior for normal markers.

## Which issue(s) this PR fixes

Fixes #140313

## Special notes for your reviewer

- Kept the change minimal and localized to the dependenttags analyzer.
- Preserves existing marker behavior.
- Added test coverage for value-qualified dependencies including:
  - missing dependency markers
  - incorrect marker values
  - missing marker values
  - valid `k8s:update=NoUnset` cases

## How has this been tested?

Ran:

go test ./pkg/analysis/dependenttags/...

All tests passed.


GaurangKhanderay:
Hi @everettraven @jpbetz  when you get a chance, could you please take a look at this PR? Happy to address any feedback. Thanks!

GaurangKhanderay:
Hi @everettraven  and @jpbetz  , just following up on this PR. Whenever you have some time, I'd appreciate a review. The change is small and includes tests for the new value-qualified marker dependency behavior. Happy to make any requested changes. Thanks!

itzPranshul (approved):
The implementation looks good to me. Thanks!

one nit: could you update `doc.go` to reflect that value matching is now supported

GaurangKhanderay:
I've updated `pkg/analysis/dependenttags/doc.go` to document support for value-qualified marker matching.

PR #253, opened 2026-07-15
Expand listType=set linter to disallow all non-scalars

jpbetz:
In https://github.com/kubernetes/kubernetes/pull/140293#pullrequestreview-4705355233 it was noticed that CRDs allow listType=set for not only objects, but also slices and maps.

Due to the lack of SSA support, we would like to discourage use of non-scalar sets. We intend to ban the use for the native types in Kubernetes since none of the REST API types yet use non-scalar sets.

### DIFF
--- a/pkg/analysis/ssatags/analyzer.go
+++ b/pkg/analysis/ssatags/analyzer.go
@@ -18,6 +18,7 @@ package ssatags
 import (
 	"fmt"
 	"go/ast"
+	"go/types"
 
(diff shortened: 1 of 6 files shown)


jpbetz:
cc @JoelSpeed

jpbetz:
cc @liggitt

JoelSpeed (commented):
/lgtm
/approve

Sounds reasonable to me

PR #256, opened 2026-07-23
fix: ignore well-known Kubernetes types in zero value validation

### LINKED ISSUE #249: requiredfields/optionalfields: false positive on metav1.Time and corev1.LocalObjectReference with omitzero
## Description

The `requiredfields` and `optionalfields` analyzers produce false positives for fields using well-known Kubernetes types like `metav1.Time` and `corev1.LocalObjectReference` when paired with `omitzero`.

## Example

```go
type EncryptionMigrationHistory struct {
	// startedTime is when the rotation was initiated.
	// +required
	StartedTime metav1.Time `json:"startedTime,omitzero"`
	// completionTime is when the rotation finished.
	// +optional
	CompletionTime metav1.Time `json:"completionTime,omitzero"`
}

type AESCBCKeyStatus struct {
	// secret is a reference to the Secret containing the key.
	// +required
	Secret corev1.LocalObjectReference `json:"secret,omitzero"`
}
```

## Lint output

```
requiredfields: field EncryptionMigrationHistory.StartedTime has a valid zero value ({}), but the validation is not complete (e.g. min properties/adding required fields). The field should be a pointer to allow the zero value to be set. If the zero value is not a valid use case, complete the validation and remove the pointer.

optionalfields: field EncryptionMigrationHistory.CompletionTime has a valid zero value ({}), but the validation is not complete (e.g. min properties/adding required fields). The field should be a pointer to allow the zero value to be set. If the zero value is not a valid use case, complete the validation and remove the pointer.

requiredfields: field AESCBCKeyStatus.Secret has a valid zero value ({}), but the validation is not complete (e.g. min properties/adding required fields). The field should be a pointer to allow the zero value to be set. If the zero value is not a valid use case, complete the validation and remove the pointer.
```

## Why this is a false positive

The linter inspects the external struct via Go's type system, counts non-omitted fields, finds zero (because all internal fields use `omitempty`/`omitzero`), and concludes the zero value `{}` is "valid but validation is incomplete."

However, for well-known Kubernetes types like `metav1.Time` (which wraps `time.Time`), a zero value is never a meaningful user choice — it represents the Go zero, not a valid timestamp. Making these fields pointers solely to distinguish "zero time" from "unset" is unnecessary when `omitzero` already handles omission correctly.

Similarly, `corev1.LocalObjectReference` with a single `Name` field that has `omitempty` — the zero value `{}` is not a valid object reference.

## Suggested fix

The linter should recognize well-known Kubernetes API types (e.g., `metav1.Time`, `metav1.Duration`, `corev1.LocalObjectReference`, `corev1.ObjectReference`, `resource.Quantity`) as types where the zero value is not a meaningful user input, and skip the "should be a pointer" suggestion for them. Alternatively, providing a configuration option to allowlist specific external types would also work.

## Version

```
sigs.k8s.io/kube-api-linter v0.0.0-20260518104151-5ebe05f9440b
```

Dasmat13:
Fixes #249

this fixes the false positive where requiredfields and optionalfields analyzers ask for a pointer on well known k8s types like metav1.Time when using omitzero. 

I just added a quick check to intercept these types in the external struct validation logic so they are ignored. tested locally and works good.

### DIFF
--- a/pkg/analysis/utils/testdata/src/c/external.go
+++ b/pkg/analysis/utils/testdata/src/c/external.go
@@ -1,0 +1,25 @@
+package c
+
+import (
+	"extpkg"
+	v1 "k8s.io/api/core/v1"
+)
+
+type ExternalTestStructs struct {
+	// Should flag: external struct without IsZero
+	TargetNoIsZero extpkg.NoIsZero `json:"targetNoIsZero,omitzero"` // want "zero value is valid" "validation is not complete"
+
+	// Should not flag: external struct with IsZero on value receiver
+	TargetValueRecv extpkg.ValueReceiver `json:"targetValueRecv,omitzero"` // want "zero value is valid" "validation is complete"
+
+	// Should not flag: external struct with IsZero on pointer receiver
+	TargetPtrRecv extpkg.PointerReceiver `json:"targetPtrRecv,omitzero"` // want "zero value is valid" "validation is complete"
+
+	// Should not flag: hardcoded exception
+	TargetLocalObjRef v1.LocalObjectReference `json:"targetLocalObjRef,omitzero"` // want "zero value is valid" "validation is complete"
+
+	// Type aliases
+	TargetAliasToNoIsZero AliasToNoIsZero `json:"targetAliasToNoIsZero,omitzero"` // want "zero value is valid" "validation is not complete"
+}
+
+type AliasToNoIsZero = extpkg.NoIsZero
--- a/pkg/analysis/utils/testdata/src/extpkg/ext.go
+++ b/pkg/analysis/utils/testdata/src/extpkg/ext.go
@@ -1,0 +1,15 @@
+package extpkg
+
+type NoIsZero struct {
+	Field string `json:"field,omitempty"`
+}
+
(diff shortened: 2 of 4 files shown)


JoelSpeed:
Why create exceptions rather than exploring the underlying issue? Or have we already explored the underlying issue to understand why the current tooling is not understanding that this is zero-able/not

Would you be able to add some tests to demonstrate your point?

Dasmat13:
hey @JoelSpeed, just a quick update on this. I was playing around with it more and realized my approach here actually introduces a bug.

since I'm returning true, true (meaning zero value is valid), the linter assumes the field is totally fine and stops checking for the omitzero tag. so if someone actually forgets to add omitzero to metav1.Time, the linter won't even warn them anymore 🤦‍♂️

saw @rishh09 just opened #257 which fixes the same thing but without this regression (they just skip the pointer check instead). we should probably just go with their PR since it's safer. we might just want to grab my methodSet lookup logic and put it in their PR since it handles pointer receivers a bit better. let me know what u think!

Dasmat13:
Closing this in favor of #257

PR #257, opened 2026-07-23
[utils] Handle omitzero for external struct types

### LINKED ISSUE #249: requiredfields/optionalfields: false positive on metav1.Time and corev1.LocalObjectReference with omitzero

rishh09:
Fixes #249.

The linter already handles local structs with `omitzero`, but external selector types could still fall through to the later pointer/`omitempty` checks. This keeps the existing `IsStructType` behavior and skips that later path only when `omitzero` is already present for an external struct or a type with `IsZero() bool`.

The tests cover external structs with `omitzero`, an external `IsZero()` type with and without the tag, and a local valid-zero struct to make sure its existing warning is unchanged.

Testing:
- `go test ./...`
- `go tool -modfile tools/go.mod github.com/golangci/golangci-lint/v2/cmd/golangci-lint run ./... --timeout 5m -v`
- `go tool -modfile tools/go.mod golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize -diff ./...`

### DIFF
--- a/pkg/analysis/utils/serialization/serialization_check.go
+++ b/pkg/analysis/utils/serialization/serialization_check.go
@@ -116,7 +116,7 @@ func (s *serializationCheck) Check(pass *analysis.Pass, field *ast.Field, marker
 		s.handleFieldShouldBePointer(pass, field, fieldName, isPointer, underlying, markersAccess, "should be a pointer.", qualifiedFieldName)
 		s.handleFieldShouldHaveOmitEmpty(pass, field, qualifiedFieldName, hasOmitEmpty, jsonTags)
 	case PointersPreferenceWhenRequired:
-		s.handleFieldOmitZero(pass, field, fieldName, jsonTags, underlying, hasOmitZero, hasValidZeroValue, isPointer, isStruct, markersAccess, qualifiedFieldName)
+		s.handleFieldOmitZero(pass, field, fieldName, jsonTags, underlying, hasOmitZero, hasValidZeroValue, isPointer, isStruct || utils.HasIsZeroMethod(pass, underlying), markersAccess, qualifiedFieldName)
 
 		if s.omitEmptyPolicy != OmitEmptyPolicyIgnore || hasOmitEmpty {
 			// If we require omitempty, or the field has omitempty, we can check the field properties based on it being an omitempty field.
@@ -153,6 +153,14 @@ func (s *serializationCheck) handleFieldShouldHaveOmitEmpty(pass *analysis.Pass,
 
 func (s *serializationCheck) checkFieldPropertiesWithOmitEmptyRequired(pass *analysis.Pass, field *ast.Field, fieldName string, jsonTags extractjsontags.FieldTagInfo, underlying ast.Expr, hasOmitEmpty, hasValidZeroValue, completeValidation, isPointer, isStruct bool, markersAccess markershelper.Markers, qualifiedFieldName string) {
 	switch {
+	case jsonTags.OmitZero && s.omitZeroPolicy != OmitZeroPolicyForbid && utils.IsZeroOmittableType(pass, underlying):
+		// Struct fields with omitzero already omit their zero value, so they do
+		// not also need omitempty or pointer indirection for omission.
+		return
+	case utils.IsZeroOmittableType(pass, underlying) && !hasValidZeroValue && s.omitZeroPolicy != OmitZeroPolicyForbid:
+		// OmitZero handling already reports the missing omitzero tag for this
+		// field. Do not also require omitempty.
+		return
 	case isStruct && !hasValidZeroValue && s.omitZeroPolicy != OmitZeroPolicyForbid:
 		// The struct field need not be pointer if it does not have a valid zero value.
 		return
--- a/pkg/analysis/utils/serialization/serialization_check.go
+++ b/pkg/analysis/utils/serialization/serialization_check.go
@@ -116,9 +116,13 @@ func (s *serializationCheck) Check(pass *analysis.Pass, field *ast.Field, marker
 		s.handleFieldShouldBePointer(pass, field, fieldName, isPointer, underlying, markersAccess, "should be a pointer.", qualifiedFieldName)
 		s.handleFieldShouldHaveOmitEmpty(pass, field, qualifiedFieldName, hasOmitEmpty, jsonTags)
 	case PointersPreferenceWhenRequired:
-		s.handleFieldOmitZero(pass, field, fieldName, jsonTags, underlying, hasOmitZero, hasValidZeroValue, isPointer, isStruct || utils.HasIsZeroMethod(pass, underlying), markersAccess, qualifiedFieldName)
+		s.handleFieldOmitZero(pass, field, fieldName, jsonTags, underlying, hasOmitZero, hasValidZeroValue, isPointer, isStruct, markersAccess, qualifiedFieldName)
 
 		if s.omitEmptyPolicy != OmitEmptyPolicyIgnore || hasOmitEmpty {
+			if s.isOmittedByOmitZero(pass, jsonTags, underlying) {
+				return
+			}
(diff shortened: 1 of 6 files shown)


pkg/analysis/utils/serialization/serialization_check.go
```diff
@@ -153,6 +153,14 @@ func (s *serializationCheck) handleFieldShouldHaveOmitEmpty(pass *analysis.Pass,
 
 func (s *serializationCheck) checkFieldPropertiesWithOmitEmptyRequired(pass *analysis.Pass, field *ast.Field, fieldName string, jsonTags extractjsontags.FieldTagInfo, underlying ast.Expr, hasOmitEmpty, hasValidZeroValue, completeValidation, isPointer, isStruct bool, markersAccess markershelper.Markers, qualifiedFieldName string) {
 	switch {
+	case jsonTags.OmitZero && s.omitZeroPolicy != OmitZeroPolicyForbid && utils.IsZeroOmittableType(pass, underlying):
+		// Struct fields with omitzero already omit their zero value, so they do
+		// not also need omitempty or pointer indirection for omission.
+		return
+	case utils.IsZeroOmittableType(pass, underlying) && !hasValidZeroValue && s.omitZeroPolicy != OmitZeroPolicyForbid:
```
JoelSpeed: Isn't this duplicating the case below? Since structs are ZeroOmittable?

pkg/analysis/utils/serialization/serialization_check.go
```diff
@@ -153,6 +153,14 @@ func (s *serializationCheck) handleFieldShouldHaveOmitEmpty(pass *analysis.Pass,
 
 func (s *serializationCheck) checkFieldPropertiesWithOmitEmptyRequired(pass *analysis.Pass, field *ast.Field, fieldName string, jsonTags extractjsontags.FieldTagInfo, underlying ast.Expr, hasOmitEmpty, hasValidZeroValue, completeValidation, isPointer, isStruct bool, markersAccess markershelper.Markers, qualifiedFieldName string) {
 	switch {
+	case jsonTags.OmitZero && s.omitZeroPolicy != OmitZeroPolicyForbid && utils.IsZeroOmittableType(pass, underlying):
```
JoelSpeed: Is this case inspired by another case in the switch? Why are we adding this now? Was some other test failing?

pkg/analysis/utils/serialization/testdata/src/pointers_when_required_omit_empty_suggest_fix_omit_zero_suggest_fix/structs.go
```diff
@@ -60,6 +62,15 @@ type TestStructs struct {
...
 	StructPtrWithOmittedRequiredFieldWithOmitEmpty *StructWithOmittedRequiredField `json:"structPtrWithOmittedRequiredFieldWithOmitEmpty,omitempty"` // want "field TestStructs.StructPtrWithOmittedRequiredFieldWithOmitEmpty does not allow the zero value. It must have the omitzero tag." "field TestStructs.StructPtrWithOmittedRequiredFieldWithOmitEmpty does not allow the zero value. The field does not need to be a pointer."
+
+	// ExternalStructWithOmitZero is an external struct with omitzero.
+	ExternalStructWithOmitZero externaltypes.StructType `json:"externalStructWithOmitZero,omitzero"`
+
+	// ExternalStructWithIsZeroWithOmitZero is an external struct with IsZero and omitzero.
+	ExternalStructWithIsZeroWithOmitZero externaltypes.StructTypeWithIsZero `json:"externalStructWithIsZeroWithOmitZero,omitzero"`
+
+	// ExternalStructWithIsZeroMissingOmitZero is an external struct with IsZero but no omitzero.
+	ExternalStructWithIsZeroMissingOmitZero externaltypes.StructTypeWithIsZero `json:"externalStructWithIsZeroMissingOmitZero"` // want "field TestStructs.ExternalStructWithIsZeroMissingOmitZero does not allow the zero value. It must have the omitzero tag."
```
JoelSpeed: How do we know this does not allow the zero value? Just because it has an `IsZero` doesn't mean that we can tell if the zero value is a valid user choice or not

pkg/analysis/utils/serialization/serialization_check.go
rishh09: Yeah, I see what you mean. This overlaps with the existing struct case for local structs. My intent was to cover selector/external struct types that do not go through the local AST-based isStruct path, but the condition ended up broader than it needs to be. I’ll narrow this so the existing local struct handling stays as-is.

pkg/analysis/utils/serialization/serialization_check.go
rishh09: This was mainly to avoid extra omitempty/pointer suggestions when an external type already has omitzero. I’ll make that intent more explicit in the code.

pkg/analysis/utils/serialization/testdata/src/pointers_when_required_omit_empty_suggest_fix_omit_zero_suggest_fix/structs.go
rishh09: Agree, IsZero() only means omitzero can omit the value; it doesn’t prove the zero value is invalid. I’ll remove that assumption from the test/fix.

